PR #68, opened 2019-10-11
Sync gradle version with gradle-wrapper.properties

deejay1:


deejay1:
Sure, I just didn't want to change too many things. Thanks for taking a look!

PR #137, opened 2021-05-05
Compare seconds to seconds in AbstractPool#isInactiveForMoreThan

violetagg:


simonbasle (approved):
oh my! yeah that's a pretty dumb mistake on my part. thanks @violetagg

PR #173, opened 2023-08-22
Too much resources may be created when using lazy-warmup

### LINKED ISSUE #172: `PoolMetrics.allocatedSize` reports a different allocation count when the allocator uses threads
## Expected Behavior
When using a threaded scheduler like `Schedulers.parallel()` or `Schedulers.single()`, the reported allocation size increments by one on a non-warmed pool. The pool allocates one element more than requested.

Leaving out `subscribeOn(…)` or using an immediate scheduler doesn't show the reported behavior. 

## Actual Behavior
The pool should not request additional items.

## Steps to Reproduce

```java
@Test
void reproCase() {
  Mono<Integer> allocator = Mono.just(1).subscribeOn(Schedulers.single());

  InstrumentedPool<Integer> pool = PoolBuilder.from(allocator).sizeBetween(3, 7).buildPool();
  pool.acquire().block();
  pool.acquire().block();

  assertThat(pool.metrics().allocatedSize()).isEqualTo(3);
}
```

## Your Environment

* Reactor version(s) used: Reactor Pool 1.0.1, Reactor 3.6.0-M1
* Other relevant libraries versions (eg. `netty`, ...):
* JVM version (`java -version`):
* OS and version (eg `uname -a`):

pderop:
When the reactor-pool is configured with the `sizeBetween` method using min/max values, more resources than expected may be created.

For example,  the following code will create 4 resources instead of the expected 3:

```
		Mono<Integer> allocator = Mono.just(1).subscribeOn(Schedulers.single());
		InstrumentedPool<Integer> pool = PoolBuilder.from(allocator).sizeBetween(3, 7).buildPool();
		pool.acquire().block();
		pool.acquire().block();
```

This is not a bug per se, but we can arrange to only create minimal necessary resources (3 in the above case).
The extra resource is created because when the second acquire takes place, the allocation for the two extra resources that are lazily created after the first acquire is still in progress and not yet completed, because the allocator is subscribed asynchronously using `.subscribeOn(Schedulers.single())`

As a result, we end up with four created resources, while we would expect to only have three resources:

- first resource created by the first acquire
- second and third resources created asynchronously due to lazy warmup triggered by the first acquire
- An unexpected fourth resource is created by the second acquire because the allocation of the second and third resources is still in progress at this point.

Fixes #172

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
+++ b/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
@@ -99,6 +99,9 @@ public class SimpleDequePool<POOLABLE> extends AbstractPool<POOLABLE> {
 
 	Disposable evictionTask;
 
+	// Flag used to avoid creating resources while warmup is in progress
+	volatile boolean warmupInProgress = false;
+
 	SimpleDequePool(PoolConfig<POOLABLE> poolConfig) {
 		super(poolConfig, Loggers.getLogger(SimpleDequePool.class));
 		this.idleResourceLeastRecentlyUsed = poolConfig.reuseIdleResourcesInLruOrder();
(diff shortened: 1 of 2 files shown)


reactor-pool/src/test/java/reactor/pool/SimpleDequePoolInstrumentationTest.java
```diff
@@ -374,4 +377,66 @@ void testIsInactiveForMoreThan() {
...
+		CountDownLatch latch = new CountDownLatch(2);
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		latch.await(5, TimeUnit.SECONDS);
+		assertThat(pool.metrics().allocatedSize()).isEqualTo(3);
+	}
+
+	@Test
+	void testIssue172_async_with_delay_and_more_than_min() throws InterruptedException {
+		Mono<Integer> allocator = Mono.just(1).delayElement(Duration.ofSeconds(1)).subscribeOn(Schedulers.single());
```
chemicL, line 416: AFAIU the `subscribeOn` operator is unnecessary, as the `delayElement` operator introduces asynchrony when delivering the result. The concurrency of that by default is the number of processors available.

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -99,6 +99,9 @@ public class SimpleDequePool<POOLABLE> extends AbstractPool<POOLABLE> {
 
 	Disposable evictionTask;
 
+	// Flag used to avoid creating resources while warmup is in progress
+	volatile boolean warmupInProgress = false;
```
chemicL, line 103: Would it make sense to prevent the `drainLoop()` being called concurrently instead? Let's say that > 1 concurrent acquires happen. Both increment the demand and add themselves to the `Borrower` queue. However, just one enters the `drainLoop` and delivers all necessary resources. A loop can check if more demand has been added before it exits in order to repeat the procedure and ensure every acquire is satisfied in the end. But that would guarantee no simultaneous warmup happens.

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -372,89 +375,97 @@ private void drainLoop() {
...
+								 * (warm up in sequence to primary allocation) *
+								 *=============================================*/
+								int toWarmup = permits - 1;
+								logger.debug("should warm up {} extra resources", toWarmup);
+
+								final long startWarmupIteration = clock.millis();
+								// flatMap will eagerly subscribe to the allocator from the current thread, but the concurrency
+								// can be controlled from configuration
+								final int mergeConcurrency = Math.min(poolConfig.allocationStrategy().warmupParallelism(), toWarmup + 1);
+								warmupInProgress = true;
```
chemicL, line 457: This flag does not guarantee exclusivity IMO. A stress test could be added to validate my accusation and I worry it would hold true :(

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
chemicL: Just to clarify why drainLoop can be executed concurrently: both evictInBackground and pendingOffer methods have a check like this:

				if (WIP.decrementAndGet(this) > 0) {
					drainLoop();
				}
				
And it doesn't exclude other threads from entering.

chemicL (commented):
My feedback is in the comments. I think I understand the justification for the need to not allocate more than was demanded, but I'm afraid the provided solution will not prevent excessive warmup in case of high concurrency. It would work when concurrency is limited I believe. And perhaps in a highly concurrent scenario we'd actually want to allocate more. That's why I'm on the fence and just commenting instead of approving/requesting changes.

reactor-pool/src/test/java/reactor/pool/SimpleDequePoolInstrumentationTest.java
```diff
@@ -374,4 +377,66 @@ void testIsInactiveForMoreThan() {
...
+		CountDownLatch latch = new CountDownLatch(2);
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		latch.await(5, TimeUnit.SECONDS);
+		assertThat(pool.metrics().allocatedSize()).isEqualTo(3);
+	}
+
+	@Test
+	void testIssue172_async_concurrent_warmup() throws InterruptedException {
+		Mono<Integer> allocator = Mono.just(1).subscribeOn(Schedulers.single());
```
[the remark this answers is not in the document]
pderop: I should have given more context about this PR, here is it:

in the context of r2dbc-pool, connection allocators are always subscribed to asynchronously (Schedulers.single() is used, please check [this](https://github.com/r2dbc/r2dbc-pool/blob/main/src/main/java/io/r2dbc/pool/ConnectionPool.java#L231)). Configuring the connection allocator with a scheduler like [this](https://github.com/r2dbc/r2dbc-pool/blob/main/src/main/java/io/r2dbc/pool/ConnectionPool.java#L231) will mean that the connection allocation will take place outside of the current Http EventLoop thread (in the context of Reactor Netty), but from the scheduler that is configured with the allocator (Schedulers.single()). This trick disables HttpResource colocation which is causing issues in the context of R2DBC (see [this](https://github.com/reactor/reactor-netty/issues/2781) for more info). So, here, the problem we have is that when users are configuring their allocators so they are asynchronous (like [this](https://github.com/r2dbc/r2dbc-pool/blob/main/src/main/java/io/r2dbc/pool/ConnectionPool.java#L231)), then more resources can be allocated in some situation, see this [PR's description](https://github.com/reactor/reactor-pool/pull/173#issue-1861758770).

reactor-pool/src/test/java/reactor/pool/SimpleDequePoolInstrumentationTest.java
pderop: indeed, subscribeOn is useless here, thanks.

I have removed it, please check 
https://github.com/reactor/reactor-pool/pull/173/commits/079c6357882226b98684f0e99e204cd108d7a86a

reactor-pool/src/test/java/reactor/pool/SimpleDequePoolInstrumentationTest.java
```diff
@@ -374,4 +377,66 @@ void testIsInactiveForMoreThan() {
...
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		pool.acquire().subscribe(integerPooledRef -> latch.countDown());
+		assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+		assertThat(pool.metrics().allocatedSize()).isEqualTo(4);
+	}
+
+	@Test
+	void testIssue172_async_with_delay_and_more_than_min_concurrent_warmup() throws InterruptedException {
+		Mono<Integer> allocator = Mono.just(1).delayElement(Duration.ofSeconds(1)).subscribeOn(Schedulers.single());
```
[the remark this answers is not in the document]
pderop: thanks, I have removed subscribeOn, please check https://github.com/reactor/reactor-pool/pull/173/commits/079c6357882226b98684f0e99e204cd108d7a86a

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -372,89 +375,97 @@ private void drainLoop() {
...
+								Flux.range(1, toWarmup)
+										.map(i -> warmupMono(i, toWarmup, startWarmupIteration, allocator).doOnSuccess(__ -> drain()))
+										.startWith(primary.doOnSuccess(__ -> drain()).then())
+										.flatMap(Function.identity(), mergeConcurrency, 1) // since we dont store anything the inner buffer can be simplified
+										.onErrorResume(e -> Mono.empty())
+										.subscribe(aVoid -> {
+										}, alreadyPropagatedOrLogged -> drain(), () -> {
+											warmupInProgress = false;
+											drain();
+										});
```
violetagg, line 467: On error `warmupInProgress` will continue to stay `true`. Is that intensional?

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
pderop: if I'm correct `drainLoop` can't be run concurrently, it is protected using the WIP design pattern.
However, I will need to give more thinking about your comment, thanks.

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
violetagg: I also think so

pderop: I do not see for the moment the concurrency issue, but I will give investigate this, and will create a stress test.
The flag is only set from the drainLoop and then it is reset once the subscription to the Flux.range completes, and if drainLoop is missing the update of warmupInProgress, then drain() will cause drainLoop to be called again.
However, since both of you are seeing an issue here, I'll carefully revisit all this.

There is however something that I missed: it is when the subscription to the `primary` is failing, in this case there is a problem because the flag is not reset to false.

Another PR is addressing this problem: https://github.com/reactor/reactor-pool/pull/175

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
pderop: good catch, the https://github.com/reactor/reactor-pool/pull/175 is partially addressing this problem.

pderop: warmpupInProgress should be indeed reset even if there is an error. will think about this.

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
chemicL: This is not a typical WIP pattern that is in use here. The above condition should be as the one in the `drain` method:

	void drain() {
		if (WIP.getAndIncrement(this) == 0) {
			drainLoop();
		}
	}
	
but it's not the case for `evictInBackground` and `pendingOffer`.

pderop:
closing this PR, see #172.

Issue #174, opened 2023-08-31
Incorrect handling of allocator errors during "warmup" in SimpleDequePool


k-tokarev:
## Expected Behavior
Pool resources should be correctly restored according to size based allocation strategy after allocator problems disappear

## Actual Behavior
If allocator can't allocate new resource (emits error in result of acquire) and pool of idle resources is empty then available permits decreasing by minSize-1 (permits for additional "warmup" resources don't return after main resource acquire failure!)

## Steps to Reproduce
```java
@Test
void reproCase() {
        val canAllocateResource = new AtomicBoolean(true);
        val allocator = Mono.defer(() ->
            canAllocateResource.get() ?
                Mono.just("value") :
                Mono.error(new IllegalStateException("Can't allocate"))
        );

        val pool = PoolBuilder
            .from(allocator)
            .maxPendingAcquireUnbounded()
            .sizeBetween(10, 10) // Spring Boot R2DBC connections pool default
            .buildPool();

        // New empty pool. No resources allocated yet, but has min-size (10) permits
        assertThat(pool.config().allocationStrategy().estimatePermitCount()).isEqualTo(10);
        assertThat(pool.metrics().idleSize()).isEqualTo(0);

        // Try to acquire one resource. This should trigger pool "warmup" to min-size of resources
        StepVerifier.create(pool.acquire().flatMap(PooledRef::release)).verifyComplete();
        assertThat(pool.config().allocationStrategy().estimatePermitCount()).isEqualTo(0);
        assertThat(pool.metrics().idleSize()).isEqualTo(10);

        // Now allocator will return errors (simulating inaccessible DB server for R2DBC connections pool)
        canAllocateResource.set(false);

        // We have 10 allocated resources in the pool, but they are not valid anymore, so invalidate them
        StepVerifier.create(Flux.range(0, 10).concatMap(ignore -> pool.acquire().flatMap(PooledRef::invalidate)))
            .verifyComplete();
        assertThat(pool.metrics().idleSize()).isEqualTo(0);
        assertThat(pool.config().allocationStrategy().estimatePermitCount()).isEqualTo(10);

        // Now we have empty pool, so it should be warmed up again but allocator still not working
        StepVerifier.create(pool.acquire()).verifyError();
        assertThat(pool.metrics().idleSize()).isEqualTo(0);
        assertThat(pool.config().allocationStrategy().estimatePermitCount()).isEqualTo(10); // Oops!
}
```

## Possible Solution
[...]

PR #175, opened 2023-08-31
Bugfix for issue #174

k-tokarev:
In order to support my issue (https://github.com/reactor/reactor-pool/issues/174#issue-1875620073) I made this pull request with test indicating the problem and my "naive" fix for it. Project builds successfully with this fix, so I hope it don't brake anything.


k-tokarev:
Sorry for imports rearranged in this pull request. My IntelliJ IDEA did it automatically and I don't know how to disable this for one single project.

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -29,22 +40,6 @@
...
-import reactor.core.Exceptions;
-import reactor.core.Scannable;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
-import reactor.core.publisher.Operators;
-import reactor.core.scheduler.Scheduler;
-import reactor.core.scheduler.Schedulers;
-import reactor.util.Loggers;
-import reactor.util.annotation.Nullable;
-
```
pderop, line 47: can you please revert the imports in order to reduce the impact of the patch ?

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -451,7 +446,7 @@ else if (sig.isOnError()) {
 							final int mergeConcurrency = Math.min(poolConfig.allocationStrategy().warmupParallelism(), toWarmup + 1);
 							Flux.range(1, toWarmup)
 									.map(i -> warmupMono(i, toWarmup, startWarmupIteration, allocator).doOnSuccess(__ -> drain()))
-									.startWith(primary.doOnSuccess(__ -> drain()).then())
+									.startWith(primary.doOnSuccess(__ -> drain()).then().onErrorComplete())
```
pderop, line 449:
```suggestion
									.startWith(primary.doOnSuccess(__ -> drain()).onErrorComplete().then())
```

although the patch seems correct, what do you think if we append `onErrorComplete` just after the `primary.doOnSuccess(...)`, it is just a matter of taste. I think it's the same, I just tend to prefer doing that in terms of clarity ?

@chemicL , do you agree with this patch ? and are you ok to move the onErrorComplete before `.then()`(just for readability reasons) ?

reactor-pool/src/test/java/reactor/pool/AllocationFailureHandlingTest.java
```diff
@@ -0,0 +1,66 @@
...
+        StepVerifier.create(Flux.range(0, 10).concatMap(ignore -> pool.acquire().flatMap(PooledRef::invalidate)))
+            .verifyComplete();
+        assertThat(pool.metrics().idleSize()).isEqualTo(0);
+        assertThat(pool.config().allocationStrategy().estimatePermitCount()).isEqualTo(10);
+
+        // Now we have empty pool, so it should be warmed up again but allocator still not working
+        StepVerifier.create(pool.acquire()).verifyError();
+        assertThat(pool.metrics().idleSize()).isEqualTo(0);
+        assertThat(pool.config().allocationStrategy().estimatePermitCount()).isEqualTo(10); // Oops!
+    }
```
pderop, line 65: What do you think if we move the test in an existing test class, for example in `CommonPoolTest.java` ?
You could then rename your `test`method to `testIssue-174` for example ?

k-tokarev:
> Here are few feedbacks :-)
> 
> @reactor/core-team, please take a look ? thanks

I committed all proposed changes. Can you review new version of Pull Request?

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
```diff
@@ -94,7 +94,7 @@ public <T, CONF extends PoolConfig<T>> AbstractPool<T> apply(PoolBuilder<T, CONF
 			return (AbstractPool<T>) builder.idleResourceReuseOrder(isLru).buildPool();
 		}
 	}
-	
+
```
pderop, line 97: nit: can you remove the extra spaces in this line ?

k-tokarev: > nit: can you remove the extra spaces in this line ?

It was automatically removed by my IDE. Do you mean restore extra spaces just to remove this diff? I will do it using GitHub editor (I don't want to disable this useful feature in my IDE).

pderop: my bad, you are right, I was meaning restore the extra space, in order to keep the patch as minimal.

pderop (approved):
Looking good from my POV, but we need @reactor/core-team to also review, maybe they have more comments to do ?

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -451,7 +451,7 @@ else if (sig.isOnError()) {
 							final int mergeConcurrency = Math.min(poolConfig.allocationStrategy().warmupParallelism(), toWarmup + 1);
 							Flux.range(1, toWarmup)
 									.map(i -> warmupMono(i, toWarmup, startWarmupIteration, allocator).doOnSuccess(__ -> drain()))
-									.startWith(primary.doOnSuccess(__ -> drain()).then())
+									.startWith(primary.doOnSuccess(__ -> drain()).onErrorComplete().then())
```
chemicL, line 454: As was discussed yesterday, this solves the issue for now but can be further improved to decouple unused permit return from connection acquisition. Right now it is implicit what happens and a reader of the code might scratch their head for some time to understand why this is used.

chemicL (approved):
Approving but hoping for a follow-up refactor to better structure the code and individual operators' responsibilities regarding connection acquisition, error handling, vs permit obtaining and returning to the pool.

pderop:
@chemicL , thanks for the review, we'll try to  refactor all this in the next version.

@k-tokarev , thanks for the PR, which is merged.

PR #177, opened 2023-09-26
Fix time unit mixup in `InstrumentedPool.PoolMetrics#isInactiveForMoreThan` default implementation

### LINKED ISSUE #176: InstrumentedPool.PoolMetrics isInactiveForMoreThan mixes time units
The `default boolean InstrumentedPool.PoolMetrics.isInactiveForMoreThan(Duration duration);` implementation compares the _seconds_ since the last interaction with the _milliseconds_ of the provided duration.

## Expected Behavior
I would expect the method to return `true` iff the last interaction was longer ago than the provided `duration`.

## Actual Behavior
The method returns `true` iff the last interaction is longer ago than the provided `duration * 1000`, basically.

## Steps to Reproduce

```java
@Test
void reproCase() {

        long inactiveSeconds = 5;
        var metrics = new InstrumentedPool.PoolMetrics() {

            @Override
            public int acquiredSize() {
                return 0;
            }

            @Override
            public int allocatedSize() {
                return 0;
            }

            @Override
            public int idleSize() {
                return 0;
            }

            @Override
            public int pendingAcquireSize() {
                return 0;
            }

            @Override
            public long secondsSinceLastInteraction() {
                return inactiveSeconds;
            }

            @Override
            public int getMaxAllocatedSize() {
                return 0;
            }

            @Override
            public int getMaxPendingAcquireSize() {
                return 0;
            }
        };
        
        assertTrue(metrics.isInactiveForMoreThan(Duration.ofSeconds(1)));
}
```

## Possible Solution

Change `this.secondsSinceLastInteraction() >= duration.toMillis()` to `this.secondsSinceLastInteraction() >= Math.ceil((double) duration.toMillis() / 1000)`.

## Your Environment

* Reactor version(s) used: `main` branch
* Other relevant libraries versions (eg. `netty`, ...): ---
* JVM version (`java -version`): ---
* OS and version (eg `uname -a`): ---

freelon:
Previously the units were mixed up and milliseconds were directly compared to seconds.

Fixes #176.

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/InstrumentedPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/InstrumentedPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -113,7 +113,7 @@ public interface InstrumentedPool<POOLABLE> extends Pool<POOLABLE> {
 		 */
 		default boolean isInactiveForMoreThan(Duration duration) {
 			return acquiredSize() == 0 && idleSize() == 0 && pendingAcquireSize() == 0 && allocatedSize() == 0
-					&& secondsSinceLastInteraction() >= duration.toMillis();
+					&& !Duration.ofSeconds(secondsSinceLastInteraction()).minus(duration).isNegative();
 		}
 
 		/**
--- a/reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
+++ b/reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2023 VMware Inc. or its affiliates, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package reactor.pool;
+
+import java.time.Duration;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class InstrumentedPoolPoolMetricsTest {
+
+	@Test
+	void testIsInactiveForMoreThan() {
+		InstrumentedPool.PoolMetrics metrics = new TestPool(2);
+
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofMillis(1999))).isTrue();
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofSeconds(2))).isTrue();
+		
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofMillis(2001))).isFalse();
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofSeconds(3))).isFalse();
+	}
+
+	static class TestPool implements InstrumentedPool.PoolMetrics {
+
+		final long inactiveSeconds;
+
+		TestPool(long inactiveSeconds) {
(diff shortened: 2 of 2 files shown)


violetagg (approved):

violetagg:
@reactor/team PTAL

reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class InstrumentedPoolPoolMetricsTest {
+
+	@Test
+	void testIsInactiveForMoreThan() {
+		InstrumentedPool.PoolMetrics metrics = new TestPool(2);
+
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofMillis(1999))).isTrue();
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofSeconds(2))).isTrue();
```
chemicL, line 32: technically, 2s is not more than 2s, so I'd say this assertion should be false

reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
```diff
@@ -0,0 +1,81 @@
...
+		
+		assertThat(metrics.isInactiveForMoreThan(Duration.ofMillis(2001))).isFalse();
```
chemicL, line 34: I think testing against 1999 millis would be useful :)

reactor-pool/src/main/java/reactor/pool/InstrumentedPool.java
```diff
@@ -113,7 +113,7 @@ default long secondsSinceLastInteraction() {
 		 */
 		default boolean isInactiveForMoreThan(Duration duration) {
 			return acquiredSize() == 0 && idleSize() == 0 && pendingAcquireSize() == 0 && allocatedSize() == 0
-					&& secondsSinceLastInteraction() >= duration.toMillis();
+					&& !Duration.ofSeconds(secondsSinceLastInteraction()).minus(duration).isNegative();
```
chemicL, line 116: Negation of negation is a bit difficult to follow, how about this?

    && duration.minusSeconds(secondsSinceLastInteraction()).isNegative()

chemicL (changes requested):
Left a few comments, pretty quick to address. Thanks for the contribution!

reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
violetagg: isn't it tested on row 31?

reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
violetagg: wasn't this implemented according javadoc?

```
		/**
		 * A convenience way to check the pool is inactive, in the sense that {@link #acquiredSize()},
		 * {@link #idleSize()}, {@link #pendingAcquireSize()} and {@link #allocatedSize()} are all at zero
		 * and that the last recorded interaction with the pool ({@link #secondsSinceLastInteraction()})
		 * was more than or exactly {@code duration} ago.
		 *
		 * @return true if the pool can be considered inactive (see above), false otherwise
		 * @see #secondsSinceLastInteraction()
		 */
```

```
that the last recorded interaction with the pool ({@link #secondsSinceLastInteraction()})
		 * was more than or exactly {@code duration} ago.
```

reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
chemicL: My bad! 👓 Thanks for pointing it out :)

reactor-pool/src/test/java/reactor/pool/InstrumentedPoolPoolMetricsTest.java
chemicL: You're right. Then it's not the implementation that is at fault, but the javadoc contradicts what the words in the method name say :(
`isInactiveForMoreThan` contradicts `more than or exactly `. In this case I have no more comments. If you're ok with this @violetagg please disregard my comment.

reactor-pool/src/main/java/reactor/pool/InstrumentedPool.java
chemicL: After tracing the history with @violetagg it seems this implementation is compliant with the javadoc and the other implementation, so it makes sense to keep it consistent. Please disregard my comment.

chemicL (approved):
After clarifications my comments are now resolved.

violetagg:
@freelon Thanks for the PR

This is merged into `0.2.x` for version `0.2.12` and into `main` for version `1.0.3`

PR #178, opened 2023-11-16
Expose metrics for `pending acquire` operation latency

violetagg:
This is related to https://github.com/reactor/reactor-netty/issues/2946

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -44,6 +44,7 @@ import static reactor.pool.AbstractPool.AbstractPooledRef.STATE_INVALIDATED;
  * related classes like {@link AbstractPooledRef} or {@link Borrower}.
  *
  * @author Simon Baslé
+ * @author Violeta Georgieva
  */
 abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 												 InstrumentedPool.PoolMetrics {
@@ -125,7 +126,7 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 
 	/**
 	 * Note to implementors: stop the {@link Borrower} countdown by calling
-	 * {@link Borrower#stopPendingCountdown()} as soon as it is known that a resource is
+	 * {@link Borrower#stopPendingCountdown(boolean)} as soon as it is known that a resource is
 	 * available or is in the process of being allocated.
 	 */
 	abstract void doAcquire(Borrower<POOLABLE> borrower);
@@ -382,6 +383,7 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 	 * an {@link AbstractPool}.
 	 *
 	 * @author Simon Baslé
+	 * @author Violeta Georgieva
 	 */
 	static final class Borrower<POOLABLE> extends AtomicBoolean implements Scannable, Subscription, Runnable  {
 
@@ -391,6 +393,7 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 		final AbstractPool<POOLABLE> pool;
 		final Duration pendingAcquireTimeout;
 
+		long pendingAcquireStart;
 		Disposable timeoutTask;
 
 		Borrower(CoreSubscriber<? super AbstractPooledRef<POOLABLE>> actual,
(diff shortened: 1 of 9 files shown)


reactor-pool/src/main/java/reactor/pool/AbstractPool.java
```diff
@@ -434,15 +440,22 @@ public void request(long n) {
 		/**
 		 * Stop the countdown started when calling {@link AbstractPool#doAcquire(Borrower)}.
 		 */
-		void stopPendingCountdown() {
+		void stopPendingCountdown(boolean success) {
+			if (!timeoutTask.isDisposed()) {
```
chemicL, line 444: Assuming timeoutTask was disposed, what should happen with the pending metric?

violetagg: Isn't the pending metric per connection pool?

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
```diff
@@ -2680,4 +2681,50 @@ void testIssue_174(PoolStyle style) {
...
+				.verify(Duration.ofSeconds(1)); //error
+
+		pooledRef.release().block(Duration.ofSeconds(1));
+
+		assertThat(recorder.getPendingTotalCount())
+				.as("total pending")
+				.isEqualTo(2);
+		assertThat(recorder.getPendingSuccessCount())
+				.as("pending success")
+				.isEqualTo(1)
```
chemicL, line 2713: The above comments say "success" twice, perhaps it makes sense to clarify a bit which one triggers the successful pending increment.

chemicL (approved):
Minor comments / questions, so approving.

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
violetagg: OK I'll add to the comment that the second one is pending. Thanks.

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
chemicL: Just trying to understand - so if a borrower was just timed out, its result can be ignored, while the acquisition will actually account for recording the metric. Do I understand it right?

violetagg: `!timeoutTask.isDisposed()` this check is to ensure that if you invoke this method twice for one and the same `Borrower` you will not record twice.

violetagg: I mean `timeoutTask` is disposed here in this method at the end of the method, so we need to guarantee that we record just once.

violetagg: If the task timeout the recording will happen in `run` method. Is that what you had in mind?

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
violetagg: comments extended

OlegDokuka (approved):

pderop (approved):

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -366,7 +366,7 @@ private void drainLoop() {
 						borrower.fail(new PoolShutdownException());
 						return;
 					}
-					borrower.stopPendingCountdown();
+					borrower.stopPendingCountdown(true);
```
chemicL, line 369: This is actually followed by `borrower.deliver` which calls `stopPendingCountdown`. Why not rely on the one in `deliver` then? Is it because of the re-scheduling onto the `acquisitionScheduler`? Perhaps there should be an abstraction for signalling that pending state is over and what the next step is so that it is encapsulated? It feels unwieldy that the logic is spread over two classes and feels error-prone. Shouldn't this actually cancel the timeout timer instead?

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
chemicL: I now see what is happening. In another comment I expressed my concern about it. Now my question would be - is there no race here? Is the timeoutTask delivered serially with the acquisition or does it run on potentially another thread? In such case, perhaps the flip of the Borrower's AtomicBoolean state would be the trigger to account for either success or failure in the transitioning from pending state?

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
```diff
@@ -409,6 +412,8 @@ Context currentContext() {
 		@Override
 		public void run() {
 			if (Borrower.this.compareAndSet(false, true)) {
+				// this is failure, a timeout was observed
+				pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
```
chemicL, line 416:
```suggestion
				stopPendingCountdown(false);
```

chemicL: Otherwise, the `timeoutTask` is not disposed

chemicL (changes requested):
Noticed an improvement to make.

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
violetagg: nice catch, thanks

chemicL (approved):

violetagg:
Thank you all for the review!

PR #171, opened 2023-06-22
Optimize pool warmup

pderop:
**Motivations:**

From https://github.com/reactor/reactor-netty/issues/2781 issue, there is a performance problem when 
using r2dbc drivers configured with spring.r2dbc.pool.initial-size  > 1.

For example, with `spring.r2dbc.pool.initial-size=4`, and `spring.r2dbc.pool.max-size=10`, the first four db connections acquired during warmup will actually use the same TcpResource event loop thread. 

Worsed situation is when initial-size == max-size, in this case all db connections will actually use the same TcpResource event loop thread.

This is related to Reactor Netty TcpResources, which are by default colocated.

While it's possible to create custom LoopResources in Reactor Netty without colocation, the #2781 issue requests for the ability to reuse the shared TcpResource EventLoopGroup without colocation instead of creating a custom LoopResource without coloc for each drivers, like Mardiadb, Postgres.

whilst this change request is making sense, I tend to think that before doing this in reactor-netty, it might be worth to address the problem in reactor pool: indeed the primary root cause of the problem is that during pool warmups (either when calling `pool.warmup()`, or when `pool.initial-size is > 1`), the reactor-pool uses some Reactor operators like `Flux.concat`  which sequentially subscribe to the first source then wait for it to complete before subscribing to the next. This leads to a situation where  during warmup, a first DB connection is created from a random TcpResource EventLoop thread, and once this first DB Connection completes (from the TcpResource EventLoop thread), then the pool will then acquire another DB Connection ... but from the TcpResource thread ; so in this case, colocation will then reuse the same TcpResource event loop thread for all remaining DB connections to be acquired during warmup process.

**Modifications:**

1) In SimpleDequeuePool.warmup(), at the end of the method, replace ` Flux.concat` by `Flux.merge`, because unlike Flux.concat, Flux.merge eagerly subscribes to **all** sources from the current thread, and this will then ensure that we won't allocate a DB connection from one of the TcpResource event loops, so this is good because colocation will not happen

2) in SimpleDequeuePool.drainloop(), at the end of the method, replace:

```
		Flux<Void> warmupFlux = Flux.range(1, toWarmup)
				.flatMap(i -> warmupMono(i, toWarmup, startWarmupIteration, allocator));

		primary.onErrorResume(e -> Mono.empty())
				.thenMany(warmupFlux)
				.subscribe(aVoid -> { }, alreadyPropagatedOrLogged -> drain(), this::drain);
```
by:

```
[...]

PR #181, opened 2023-12-21
Warmup throws IllegalArgumentException in case no permits are available

### LINKED ISSUE #180: Warmup procedure fails if warmups were found
## Expected Behavior
When a warmup method is run, but no warmups to run were found it should not do anything.

## Actual Behavior
When a warmup method is run, but no warmups to run were found SimpleDequePool picks mergeConcurrency as lowest number of configured warmupParallelism and length of found warmups. Found warmups is 0, thus this values is passed further to Flux.merge and results in error from FluxFlatMap.
`java.lang.IllegalArgumentException: maxConcurrency > 0 required but it was 0`

## Steps to Reproduce
We have configured scheduled warmup with our custom find out idle timeout that caused pool to increase our response time. Thus we are running warmup once per X second, but it's not always needed as in high traffic this just simply should be ignored. First run after app starts is fine as then maxConcurrency is 1, but later it's mostly 0 and we just have constant errors.

## Possible Solution
Add check before executing warmups if there is anything to run to avoid unnecessary call for warmup with invalid configuration.

## Your Environment
Spring Boot 3 App with Netty server and reactor-pool 1.0.3.

* Reactor version(s) used: 1.0.3
* Other relevant libraries versions (eg. `netty`, ...): netty, Spring Boot 3
* JVM version (`java -version`): 17

pderop:
This fixes a regression that comes from #171, from 1.0.1 version: 

since the pool warmup method allows to be called at anytime, then if no more permits are available from the allocation strategy, the Flux.merge method is then called with a `mergeConcurrency`pameter that is set to `0`, causing the following exception:

```
maxConcurrency > 0 required but it was 0
java.lang.IllegalArgumentException: maxConcurrency > 0 required but it was 0
	at reactor.core.publisher.FluxFlatMap.<init>(FluxFlatMap.java:74)
	at reactor.core.publisher.Flux.merge(Flux.java:1406)
	at reactor.core.publisher.Flux.merge(Flux.java:1375)
```

Added a check in warmup to immediately return in case no permits are available.
Also added a test case.

Fixes #180

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
+++ b/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
@@ -230,6 +230,9 @@ public class SimpleDequePool<POOLABLE> extends AbstractPool<POOLABLE> {
 				recordInteractionTimestamp();
 				int initSize = poolConfig.allocationStrategy()
 				                         .getPermits(0);
+				if (initSize <= 0) {
+					return Mono.just(0);
+				}
 				@SuppressWarnings({ "unchecked", "rawtypes" }) //rawtypes added since javac actually complains
 				Mono<POOLABLE>[] allWarmups = new Mono[initSize];
 				for (int i = 0; i < initSize; i++) {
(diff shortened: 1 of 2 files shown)


pderop:
@reactor/core-team , can you please have a look ?

chemicL (approved):

pderop:
@chemicL, thank you for the review.

PR #214, opened 2024-05-14
Fix recordPendingSuccess/FailureAndLatency not recorded without timeout

HeartPattern:
recordPendingSuccess/FailureAndLatency seems not recorded properly without timeout.

If we call acquire() without timeout, timeoutTask remain as TIMEOUT_DISPOSED since AbstractPool#430 check if pendingAcquireTimeout is non-zero.

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
@@ -443,22 +443,17 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 		 * Stop the countdown started when calling {@link AbstractPool#doAcquire(Borrower)}.
 		 */
 		void stopPendingCountdown(boolean success) {
-			if (timeoutTask == null) {
-				return;
+			if (pendingAcquireStart > 0) {
+				if (success) {
+					pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
+				} else {
+					pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
+				}
+
+				pendingAcquireStart = 0;
 			}
 
 			timeoutTask.dispose();
-			timeoutTask = null;
-
-			if (pendingAcquireStart == 0) {
-				return;
-			}
-
-			if (success) {
-				pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
-			} else {
-				pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
-			}
 		}
 
 		@Override
--- a/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2023 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2024 VMware Inc. or its affiliates, All Rights Reserved.
  *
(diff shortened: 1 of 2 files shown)


reactor-pool/src/main/java/reactor/pool/AbstractPool.java
```diff
@@ -441,14 +441,18 @@ public void request(long n) {
...
-				}
+			if (timeoutTask == null) {
+				return;
 			}
+
+            if (success) {
+                pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
+            } else {
+                pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
+            }
```
violetagg, line 452: nit: Please use `tabs` not `spaces`

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
```diff
@@ -441,14 +441,18 @@ public void request(long n) {
...
-					pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
-				} else {
-					pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
-				}
+			if (timeoutTask == null) {
+				return;
 			}
+
+            if (success) {
+                pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
```
violetagg, line 449: When the timeout is not specified how/where is `pendingAcquireStart` initialised?

violetagg (changes requested):
@HeartPattern Thanks for the PR!

Please update the copyright end year with the following command `./gradlew :spotlessApply`

Can you please provide test similar to reactor.pool.CommonPoolTest#recordsPendingCountAndLatencies

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
HeartPattern: Fix pendingAcquireStart to initialized when timeout is not specified

violetagg (approved):
@reactor/team PTAL

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
```diff
@@ -441,14 +443,22 @@ public void request(long n) {
...
+
+			if (pendingAcquireStart == 0) {
+				return;
+			}
+
+			if (success) {
+				pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
+			} else {
+				pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
+			}
```
pderop, line 461: @HeartPattern , first, thanks for your contribution;

I think it's ok, but I wonder if we can reduce the size of the changes done in the `stopPendingCountdown` method.
What do you think about doing the following in the stopPendingCountdown method ?

```suggestion
			if (pendingAcquireStart > 0) {
				if (success) {
					pool.metricsRecorder.recordPendingSuccessAndLatency(pool.clock.millis() - pendingAcquireStart);
				} else {
					pool.metricsRecorder.recordPendingFailureAndLatency(pool.clock.millis() - pendingAcquireStart);
				}
				pendingAcquireStart = 0;
			}
			timeoutTask.dispose();
```

HeartPattern: That seems great. Added new commit.

pderop (approved):

pderop:
@violetagg,  can you re-review this ?

violetagg (approved):
nice

violetagg:
I back ported this fix to `1.0.x` branch. The fix will be available in `1.0.6` and `1.1.0-M3` releases.

PR #221, opened 2024-06-13
When applying `pendingAcquireTimeout`, check for current pending Borrowers

### LINKED ISSUE #219: pendingacquiretimeout is not working as expected
It seems that pendingacquiretimeout is not working as expected. 
I expected that it is timeout to acquire open connection from pool. So I expected to see timeouts occurred in my unit test in case of connection lack , but timeout is ignored and requests still wait connenction.
In attached log pendingacquiretimeout = 200ms

```
06/05/2024 19:06:40.584 [DEBUG] reactor.netty.resources.PooledConnectionProvider:pool-4-thread-4 [Loggers.java:259:debug] {} > Creating a new [webclient-pool] client pool [PoolFactory{evictionInterval=PT2S, leasingStrategy=fifo, maxConnections=1, maxIdleTime=300000, maxLifeTime=-1, metricsEnabled=false, pendingAcquireMaxCount=10, **_pendingAcquireTimeout=200_**}] for [localhost:8000]
```

and some of requests got into pending queue

```
Created a new pooled channel, now: 0 active connections, 0 inactive connections and **_5 pending acquire requests._**
```

But inspite of timeout should occur for some requests, its still live in queue and wait connection

Logs and test are attached

[log.txt](https://github.com/user-attachments/files/15593704/log.txt)
[ReactorNettyPoolWebfluxTests2.txt](https://github.com/user-attachments/files/15593705/ReactorNettyPoolWebfluxTests2.txt)

dependencies

<dependency>
    <groupId>io.projectreactor.netty</groupId>
    <artifactId>reactor-netty-http</artifactId>
    <version>1.0.33</version>
</dependency>
<dependency>
    <groupId>io.projectreactor.netty</groupId>
    <artifactId>reactor-netty-core</artifactId>
    <version>1.0.33</version>
</dependency>

java version 1.8
OS windows 11

violetagg:
Fixes #219

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
@@ -122,6 +122,18 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 
 	// == common methods to interact with idle/pending queues ==
 
+	void applyPendingAcquireTimeout(Borrower<POOLABLE> borrower) {
+		boolean noIdle = idleSize() == 0;
+		boolean noPermits = poolConfig.allocationStrategy().estimatePermitCount() == 0;
+
+		if (noIdle && noPermits) {
+			borrower.pendingAcquireStart = clock.millis();
+			if (!borrower.pendingAcquireTimeout.isZero()) {
+				borrower.timeoutTask = config().pendingAcquireTimer().apply(borrower, borrower.pendingAcquireTimeout);
+			}
+		}
+	}
+
 	abstract boolean elementOffer(POOLABLE element); //used in tests
 
 	/**
@@ -423,16 +435,8 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 		public void request(long n) {
 			if (Operators.validate(n)) {
 				//start the countdown
+				pool.applyPendingAcquireTimeout(this);
 
-				boolean noIdle = pool.idleSize() == 0;
-				boolean noPermits = pool.poolConfig.allocationStrategy().estimatePermitCount() == 0;
-
-				if (noIdle && noPermits) {
-					pendingAcquireStart = pool.clock.millis();
-					if (!pendingAcquireTimeout.isZero()) {
-						timeoutTask = this.pool.config().pendingAcquireTimer().apply(this, pendingAcquireTimeout);
-					}
-				}
 				//doAcquire should interrupt the countdown if there is either an available
 				//resource or the pool can allocate one
 				pool.doAcquire(this);
--- a/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
@@ -122,18 +122,6 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 
 	// == common methods to interact with idle/pending queues ==
 
-	void applyPendingAcquireTimeout(Borrower<POOLABLE> borrower) {
-		boolean noIdle = idleSize() == 0;
-		boolean noPermits = poolConfig.allocationStrategy().estimatePermitCount() == 0;
-
-		if (noIdle && noPermits) {
-			borrower.pendingAcquireStart = clock.millis();
-			if (!borrower.pendingAcquireTimeout.isZero()) {
-				borrower.timeoutTask = config().pendingAcquireTimer().apply(borrower, borrower.pendingAcquireTimeout);
-			}
-		}
-	}
-
 	abstract boolean elementOffer(POOLABLE element); //used in tests
 
 	/**
@@ -434,11 +422,7 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>,
 		@Override
 		public void request(long n) {
 			if (Operators.validate(n)) {
-				//start the countdown
-				pool.applyPendingAcquireTimeout(this);
-
-				//doAcquire should interrupt the countdown if there is either an available
-				//resource or the pool can allocate one
+				// doAcquire will check for acquire timeout
 				pool.doAcquire(this);
 			}
 		}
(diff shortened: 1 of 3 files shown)


reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
```diff
@@ -1117,6 +1119,31 @@ void discardCloseableWhenCloseFailureLogs(PoolStyle configAdjuster) {
...
+	@MethodSource("allPools")
+	void pendingTimeout(PoolStyle configAdjuster) throws Exception {
+		PoolBuilder<String, ?> builder = PoolBuilder
+				.from(Mono.just("pendingTimeout"))
+				.sizeBetween(0, 1)
+				.maxPendingAcquire(10);
+		AbstractPool<String> pool = configAdjuster.apply(builder);
+
+		CountDownLatch latch = new CountDownLatch(3);
+		ExecutorService executorService = Executors.newFixedThreadPool(20);
```
chemicL, line 1132: Make sure to shutdown the thread pool.

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
```diff
@@ -1117,6 +1119,31 @@ void discardCloseableWhenCloseFailureLogs(PoolStyle configAdjuster) {
 		}
 	}
 
+	@ParameterizedTestWithName
+	@MethodSource("allPools")
+	void pendingTimeout(PoolStyle configAdjuster) throws Exception {
+		PoolBuilder<String, ?> builder = PoolBuilder
+				.from(Mono.just("pendingTimeout"))
```
chemicL, line 1126: The "pendingTimeout" is confusing here -> consider calling it "pooled" for instance.

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
```diff
@@ -1117,6 +1119,31 @@ void discardCloseableWhenCloseFailureLogs(PoolStyle configAdjuster) {
...
+		for (int i = 0; i < completableFutures.length; i++) {
+			completableFutures[i] = CompletableFuture.runAsync(
+					() -> pool.acquire(Duration.ofMillis(10))
+							.doOnError(t -> latch.countDown())
+							.onErrorResume(PoolAcquireTimeoutException.class, t -> Mono.empty())
+							.block(),
+					executorService);
+		}
+
+		Mono.fromCompletionStage(CompletableFuture.allOf(completableFutures)).block(Duration.ofSeconds(5));
```
chemicL, line 1143: The mix of `CompletionStage`/`CompletableFuture` APIs makes the test a bit harder to follow. Consider picking a single API for the infrastructure around the actual SUT. With reactor the parallelization can be also achieved with a `flatMap`, while with `CompletableFuture.allOf()` you can call `join()` on them.

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
```diff
@@ -122,6 +122,18 @@ public boolean isInactiveForMoreThan(Duration duration) {
 
 	// == common methods to interact with idle/pending queues ==
 
+	void applyPendingAcquireTimeout(Borrower<POOLABLE> borrower) {
+		boolean noIdle = idleSize() == 0;
+		boolean noPermits = poolConfig.allocationStrategy().estimatePermitCount() == 0;
```
chemicL, line 127: This bit of code is not synchronized in any way, right? So it can happen that following this method call, when no timeout was launched, the borrower will still not get granted a resource and will potentially exceed the timeout waiting for a resource. Am I missing some context that guarantees atomicity of this operation or exclusiveness?

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -581,6 +585,17 @@ void pendingOffer(Borrower<POOLABLE> pending) {
...
+		int estimatePermitCount = poolConfig.allocationStrategy().estimatePermitCount();
+
+		if (idle + estimatePermitCount < postOffer) {
+			pending.pendingAcquireStart = clock.millis();
+			if (!pending.pendingAcquireTimeout.isZero()) {
+				pending.timeoutTask = config().pendingAcquireTimer().apply(pending, pending.pendingAcquireTimeout);
+			}
+		}
+
+		int maxPending = poolConfig.maxPending();
```
chemicL, line 598: Analogously to the comment in `AbstractPool`, I am wondering if the conditions leading to a `timeoutTask` being started can hold true even after the variables are assigned and when compared fail the test.

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
violetagg: done

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
violetagg: done

reactor-pool/src/test/java/reactor/pool/CommonPoolTest.java
violetagg: used `join()`

reactor-pool/src/main/java/reactor/pool/AbstractPool.java
violetagg: removed that code

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
violetagg: If you ask whether we can grand a resource and at the same time the timeout happens - yes this can happen with the current implementation.

chemicL: I'm wondering if the check in line 591 fails, but then a context switch happens and now it would hold true, while the timeoutTask has not been started. Then the pending acquisition is not timed out.

violetagg: yeah - this is `best effort`

violetagg:
@reactor/team Can you review the latest changes?

PR #256, opened 2024-12-13
Cancel allocator after borrower is canceled.

qnnn:
I recently raised an issue ([#3531](https://github.com/reactor/reactor-netty/issues/3531)) in reactor-netty related to this PR. I found that in reactor-pool, when a Borrower is canceled, the allocator is not canceled as well. It seems appropriate to cancel the allocator afterward.  Therefore, I’m attempting to submit this PR. If there are any problems, could you please provide some guidance?

Modification: 
Replace the unit test allocatedReleasedIfBorrowerCancelled with cancelAllocatorIfBorrowerCancelled.

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
+++ b/reactor-pool/src/main/java/reactor/pool/AbstractPool.java
@@ -395,6 +395,8 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>, Ins
 		long pendingAcquireStart;
 		Disposable timeoutTask;
 
+		Disposable.Composite cancelTask = Disposables.composite();
+
 		Borrower(CoreSubscriber<? super AbstractPooledRef<POOLABLE>> actual,
 				AbstractPool<POOLABLE> pool,
 				Duration pendingAcquireTimeout) {
@@ -445,6 +447,7 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>, Ins
 		@Override
 		public void cancel() {
 			set(true);
+			cancelTask.dispose();
 			pool.cancelAcquire(this);
 			stopPendingCountdown(true); // this is not failure, the subscription was canceled
 		}
@@ -478,6 +481,10 @@ abstract class AbstractPool<POOLABLE> implements InstrumentedPool<POOLABLE>, Ins
 			if (!get()) {
 				actual.onError(error);
 			}
+		}
+
+		void onCancel(Disposable disposable){
+			cancelTask.add(disposable);
 		}
 
 		@Override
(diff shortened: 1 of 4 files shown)


violetagg:
@chemicL The current behaviour is that once the resource creation is started, it is not stopped in case Borrower is cancelled in the meantime. Instead we deliver the resource to the Borrower(cancelled) and it just releases the resource to the pool so that the resource can be reused by other Borrower.
This PR changes the current behaviour and stops the creation of the resource. Wdyt?
IMO we should not change the current behaviour. Stopping the creation of the resource in the middle might be dangerous.

violetagg:
Meantime I added two checks for cancelled Borrower so that we don't start the acquisition in this cases
#263 and #265

qnnn:
Is it necessary to also consider the cancel scenario on the allocator to ensure it can be used more broadly? Apart from the cancel triggered by the changes in this modification, the allocator's own implementation may also trigger a cancel (such as timeout).

chemicL:
I agree with @violetagg, this behaviour change is undesired. An acquisition is independent of the borrower. If a particular borrower gets cancelled, the resource can still be reused by another borrower, that's the essence of pooling these resources. In the scenario you speak of - when the allocator is itself cancelled, please provide a reproducible scenario, perhaps in a new issue?

violetagg:
@chemicL @qnnn Let me close this one.

qnnn:
>  In the other scenario you speak of - when the allocator is itself cancelled, please provide a reproducible scenario, perhaps in a new issue?

@chemicL This is not an issue that currently exists, but I was thinking that when writing the allocator, there might be timeouts or exceptions that require considering discarding or destroying resources. Should resource cleanup during cancellation also be considered? This could potentially benefit sequence reuse.

> An acquisition is independent of the borrower.

I initially thought this should be a coherent action.

violetagg:
@qnnn IMO the allocator needs to take care of this (it should be able to do this as it is Publisher), so I don't see the relation between Borrower been cancelled and allocator been cancelled

qnnn:
@violetagg Thanks, I got it🫡. In the reactor-netty issue, was the callback registration for PooledConnectionInitializer and ClientTransportSubscriber missed?

qnnn:
@violetagg Oh, thank you so much! The documentation is very clear. Maybe I misunderstood from the beginning; I thought this process could be cancelled. Apologies for wasting so much of your time with support.

> In fact, Reactor Netty doesn't encourage usage of timeout operator within HTTP client because it is too broad.

Indeed, when using Spring Cloud Gateway, the responseTimeout is implemented with the timeout operator in NettyRoutingFilter. When the timeout is triggered, it's not always clear which step caused the issue (the request may have been sent or not). This doesn't mean there’s a problem with it; it still remains a good implementation for accurately controlling the request duration.

Issue #273, opened 2025-02-07
Migrate to JSpecify annotations for nullability constraints


violetagg:
https://github.com/jspecify/jspecify

PR #277, opened 2025-02-12
Annotate with `Nullable`, fields that can become null

violetagg:
Related to #273

### DIFF
--- a/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
+++ b/reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
@@ -831,7 +831,8 @@ public class SimpleDequePool<POOLABLE> extends AbstractPool<POOLABLE> {
 
 		@Override
 		public void request(long l) {
-			if (upstream != null && Operators.validate(l)) {
+			assert upstream != null;
+			if (Operators.validate(l)) {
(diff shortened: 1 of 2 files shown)


reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
```diff
@@ -829,7 +831,7 @@ public void onSubscribe(Subscription s) {
 
 		@Override
 		public void request(long l) {
-			if (Operators.validate(l)) {
+			if (upstream != null && Operators.validate(l)) {
```
chemicL, line 834: We know this can't be called unless the Subscriber was offered a reference of `this`. And that would only happen on line 828. One line above the value is assigned and on line 825 it's validated to be non-null. Therefore, perhaps we just need `assert upstream != null` and avoid the runtime null check?

chemicL (changes requested):

reactor-pool/src/main/java/reactor/pool/SimpleDequePool.java
violetagg: fixed

chemicL (approved):
LGTM!

violetagg:

